PR #3, opened 2011-03-02
Fix invalid HTML

dave1010:
Octocat likes valid markup.

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,7 +30,7 @@
 
 <body>
 
-<img src="forkit.gif" width="384" height="236" id="octocat">
+<img src="forkit.gif" width="384" height="236" id="octocat" alt="" />
 
 <p>
 	Fork me? No, fork you!


cqr:
Hey actually this is not valid HTML, it's valid XHTML. In HTML, the &lt;img&gt; tag doesn't close.

You're still right, but I would rather see them move to &lt;!DOCTYPE html&gt;

dave1010:
Very true. Though you could argue that XHTML is a type of HTML. I was just validating against the existing doctype. Changing it to `<!DOCTYPE html>` is a good idea.

Monoverde888:
uh

ivan-developer-01 (approved):

ivan-developer-01 (approved):
URL unshort

GreenePetal:
..

creeper12356:
wow First PR..

PR #17, opened 2011-04-27
Rainbow show

fletchgqc:
If a block of text has a style of display:none, does it use monitor energy?


fletchgqc:
Fantastic idea, really.

PR #26, opened 2011-06-13
might as well show the rainbow message

kpassapk:


zhuowei:
Currently, the message is shown when a person enters (spoiler!) the Konami code. I thought that it would have been a nice touch to add this little easter egg.

cameronmcefee:
As this is a demo repo, I want to keep the clutter to a minimum. The Konami code bit is a fun laugh for people that are crazy enough to just try it, or who look through the code and see it there, but I don't want them to be presented with it initially.

I do, however, thank you for contributing!

PR #34, opened 2011-07-01
test

makerdude:
test


chrisyco:
I don't think this should be merged -- the readme's fine the way it is. :)

PR #104, opened 2011-10-27
Just give a try of pull request

jianinz:
Just give a try of pull request, doesn't need to be approved or whatsoever.


jianinz:
First try to do pull request

PR #504, opened 2012-10-04
Chopsticks

lukel97:
How could we forget them?


README
```diff
@@ -1 +1 @@
-All that's missing is the fork. Heh.
\ No newline at end of file
+All that's missing is the fork. And chopsticks. Heh.
```
BKJackson, line 1: Where is the salt?

PR #522, opened 2012-10-20
Improved HTML5 version

terrill:
Removed extraneous markup (not required under html5) to make this file
more lean


terrill:
Of course this is just a test, but it's also a legitimate improvement.

PR #525, opened 2012-10-22
Mybranch

r-matous:
next round of playing with


r-matous:
just playing with => close

PR #545, opened 2012-11-05
Background and Border

Halum:
Just added a little background and border to image. Its cool isn't it  ?

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,7 +53,7 @@
 <img src="forkit.gif" id="octocat" alt="" />
 
 <p>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -15,7 +15,13 @@
       display: block;
       width:384px;
       margin: 50px auto;
+	  border: groove;
     }
+	
+	body{
+		background-color:#CFC;
```
Shockper, line 22: I don't like that color! D:

(just testing)

PR #560, opened 2012-11-12
Mybranch 123456

r-matous:
This is my next pull  request


r-matous:
Hmm, not such a big improvement!

r-matous:
Anyway safe changes.

index.html
```diff
@@ -53,7 +53,7 @@
```
r-matous: jhgjhjh

PR #619, opened 2012-12-25
Create test

tiger002:
test


EsiKhoob:
there is no problem , I think

PR #637, opened 2013-01-07
Switching cats for the - far superior - dogs

Sleavely:


index.html
```diff
@@ -28,7 +28,6 @@
       right: 10px;
       top: 10px;
       box-shadow: -5px 10px 15px gray;
-      -moz-box-shadow: -5px 10px 15px gray;
```
Sleavely, line 31: Instead of removing Mozilla's, shouldn't we add -webkit-, -ms-, and -o- prefixes?

Sleavely, line 31: No. That wouldn't be fair to Blackberry.

Issue #712, opened 2013-02-10
Your html is so broken!


PureMunky:
The title says it all...your HTML isn't what it should be, demerit.

Now what is the international symbol for sarcasm? Oh, there isn't one?

PR #713, opened 2013-02-10
Trying to attach my updated code to an issue.

PureMunky:
Do I do it like this? and then have the option later?


PureMunky:
This fixes Issue #712

PureMunky:
Great, looks good.

PR #756, opened 2013-02-27
this is a test

jntme:
and so is this.


teeeest.txt
```diff
@@ -0,0 +1 @@
+asdfas cccct test
```
jntme, line 1: testcomment

PR #793, opened 2013-03-22
Git-Knatkin

kingclowndean:
Spoon-

Knife


kingclowndean:
(Saved Fri Apr 05 01:43:51 2013)since the creation of the group you artemis3 warned that distort everything was in this group, no land, no activity ! Now you can see that we were telling the truth, diana00000 account has been deleted and no one can run this group! it is completely useless, you have been used by that person to hurt! then leave it! thank you

PR #903, opened 2013-05-09
Update README

hubot:
I added an exclamation point. Show some enthusiasm, will you?!


cnsxbjmluo:
Ripple as an open payment network, through the payment network can transfer any kind of currency, in fact, it already has most of the functions of the currency, a new virtual currency.I think that the ripple should strengthen more open and convenient payment and conversion functions can always like a stock exchange and trading.For example, a variety of currency exchange between the various companies and individuals to pay these companies and individuals may not know.
Above using Google Translator.

Issue #909, opened 2013-05-13
Yet another test issue


marianes:
# Demo issue.

Well, everybody must learn how to do things in the right order.
- [x] Task 1
- [x] Task 2
- [x] Task 3

PR #910, opened 2013-05-13
Fix for issue #909 pull request

marianes:
all tasks completed


marianes:
Why requesting a pull creates new issue?

PR #931, opened 2013-05-22
Fixed bug in README file causing build failure

norrislee:
If the title made any sense, you have issues.


AndreiDuma:
Haha, good one ;)

PR #986, opened 2013-06-16
Mension Monty Python's dirty fork sketch.

cond:
It is awfully nice to have a mension about Monty Python's sketch.


garrettmoss:
BUGFIX: !! I FOUND A BUG !!
Mention is actually spelled with a "t" :)

PR #1005, opened 2013-06-23
just a try

yangweigbh:
just try the pull request feature


index.html
```diff
@@ -10,7 +10,7 @@
       margin:0px;
       padding:0px;
     }
-    
+    #hello world
```
yangweigbh, line 13: why add this line

PR #1076, opened 2013-07-22
added a note about a spoon

slauriere:
what's a fork without a spoon


index.html
```diff
@@ -55,6 +55,7 @@
 <p>
   Fork me? No, fork you!
 </p>
+<p>Push the spoon, first.</p>
```
benglin, line 58: But there's no spoon...

PR #1147, opened 2013-08-26
Update the paragraph content

mayil30:


index.html
```diff
@@ -53,6 +53,7 @@
 <img src="forkit.gif" id="octocat" alt="" />
 
 <p>
+  This is modified files.
```
mayil30, line 56: For test purpose

PR #1202, opened 2013-09-16
newComers test

camp-fire:
just for a try o(╯□╰)o


README
```diff
@@ -1 +1 @@
-All that's missing is the fork. Heh.
\ No newline at end of file
+All that's missing is the fork. Heh....Heh.
```
camp-fire: note test

camp-fire: not test too

PR #1227, opened 2013-09-25
sample pull

vwadhwa:
more about a sample pull


RandomFolder/index.html
```diff
@@ -0,0 +1,63 @@
+<!DOCTYPE html>
+
```
vwadhwa: Why is this newline here?

vwadhwa: May be due to this or ... may be due to that ... who knows?

PR #1275, opened 2013-10-12
Image only plays/loops on hover

misterjaytee:
Created static image (forkit-static.gif) using single frame of animation.

Used jquery to swap to animated gif (forkit.gif) when mouse is hovered over image; and back to static image when mouse is not over the image.


misterjaytee:
Created Pull request to test the pull process.

PR #1283, opened 2013-10-17
I'm forking this, and I'm going to make Octocat purr!

owenbrown:
I get it now. I think we needed to make like a comment or something. The equivalent of a read-me. Maybe that's how you start all new forks.


owenbrown:
And I need a comment!

PR #1371, opened 2013-11-21
please review this update

quangnh:
I added one spoon
Then I added one more spoon


quangnh:
I'm not sure it is right

PR #1542, opened 2014-01-29
Added meta tags

NewellServ:
meta keywords and meta description tags for the omnipresent search
engine bots.


ghost:
Thanks Nazca

PR #1991, opened 2014-04-25
Changed the default content text.

lisa-usgs:
Longer description of change goes here.


lisa-usgs:
Just playing and learning more about github.

PR #2101, opened 2014-05-16
Some suggested clean up

xpucadad:
These changes were done primarily as practice to learn the appropriate protocol for contributing.
- Commented out the &lt;img> tag since the referenced image isn’t in the
  project. Maybe I'll create an issue to try and locate the gif.
- Suggested that we be nice. This change is really only the beginning. For the next the struck out code should be removed.


xpucadad:
I know this will never get pulled, so I'm just going to close it.

PR #2483, opened 2014-07-15
my change

nina124:
from nina 7.15


nina124:
discussion?

nina124:
try again

PR #2977, opened 2014-09-01
html body default to fufill the window

beswarm:
my first pull request in github


beswarm:
this is really wonderful

PR #3166, opened 2014-09-19
Update index.html

lizha:
Following new HTML5 structure


lizha:
Update index.html using standard structure of HTML5

PR #3302, opened 2014-09-30
Summary1

kokob:
description1


kokob:
made some changes on the css file

PR #3519, opened 2014-10-17
Update index.html

jhlopezm:
This is my propose  :O

### DIFF
--- a/index.html
+++ b/index.html
@@ -18,3 +18,5 @@
 
 </body>
 </html>
+
+Hello, I am here :)


index.html
```diff
@@ -18,3 +18,5 @@
 
 </body>
 </html>
+
+Hello, I am here :)
```
c99SRS, line 22: <html>
<title> My aim </title>
<body>
i like coding. I like python and java.<br>
i here to acquire knowledge.
</body>
</html>

jhlopezm, line 22: Hello! 

I wrote in that repository because it was part of some tutorial, I thought we were free to do it. However, I like that you like coding because I also like it. I have programmed in C, C++, and I am learning Python and it would be interesting Java too. 

Success!

PR #3542, opened 2014-10-20
An extremly use(ful|less) patch.

upvas:


index.html
```diff
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
-  Fork me? Fork you, @octocat!
+  Fork me? Fork you, @octocat! Yay!
```
upvas: Yeaaaaha!

PR #3835, opened 2014-11-15
Update styles.css

nessa8611:


nessa8611:
first pull request...seeing how it works.

Devaryan:
Share the detailed  problem description

PR #3933, opened 2014-11-24
Changed a width for a test

ghost:
Testing!

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -11,7 +11,7 @@
 
 p {
   display: block;
-  width: 400px;
+  width: 500px;
   margin: 50px auto;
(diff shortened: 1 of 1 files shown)


bostoncampbell:
Thanks for this example and the explanation.  I have realized in the process that I’ve never seen the following display on my computer.  Is my browser preventing it?  Is there a setting I have on my computer that is preventing it?  I’ve never seen this view with the three “traffic lights” In the corner and I also don’t recall seeing the central buttons nor Synch.  I believe this is supposed to be in the local repository – right?  The Github doesn’t look like that in my computer’s access to that remote repository.

PR #4112, opened 2014-12-10
Add forkit.gif

bugaevc:
From https://github.com/knotdvn/fork-spoon
Used in index.html

The file was missing, resulting in an ugly _can't load that file_ icon.
image

After the change everithing looks right.


JerryChin:
Well done, the maintainer is supposed to merge it into upstream.

PR #4135, opened 2014-12-13
Update README.md

DDelbende:
Just for understanding how this works !


2774599273:
let's add a contributing file so we can work better together

2774599273:
let's add a contributing file so we can work better together

PR #4287, opened 2014-12-29
Index.html additional content

h51thompson:
After the initial paragraph, added an additional <p> element containing some more content. Gives users more options to practice editing a file.


h51thompson:
Got this far. Attempt to finish out exercise by having my contribution added to the master repository.

PR #4558, opened 2015-01-21
Guess this is the commit message...

jhubers:
At some locations it says push and pull at other places it says
synchronize. I guess it is the same... Now below it shows a mark button
with Commit to master as text. Guess have to  push that...


jhubers:
Maybe have to write a comment to be able to send a pull request?

PR #5027, opened 2015-02-28
Learn by Doing.

jfixemer:
I changed one line, why is the whole thing deleted and added.
Is it because I am on Windows?  Did my editor change CR/LF?  How do I see that?
Git first impression is weak sauce...


jfixemer:
Done learning.

PR #5241, opened 2015-03-18
My commit trying forking repo.

guo-yuan:
here is my commit in forked repo.


guo-yuan:
maybe drag someone here to pull my commit...

PR #5736, opened 2015-05-04
added a nice background color and fixed some inconsistent formatting

ahovgaard:
practicing pull requests


ahovgaard:
Much improved. Very wow.

PR #5743, opened 2015-05-05
Update README.md

Dishank-Chopra:
got it!!!!!!


Dishank-Chopra:
umm... want to make changes in window and view them in browser...

PR #5861, opened 2015-05-14
Added name to README.md

JanaThalheim:
Added name to README.md


test.md
```diff
@@ -0,0 +1 @@
+asd
```
JanaThalheim, line 1: I think that's the wrong file I changed...

PR #5904, opened 2015-05-19
Added newlines where appropriate

thoke:


thoke:
On second thought, this isn't needed ;-)

PR #6178, opened 2015-06-14
Added forkit.gif image

ThePavolC:
Project doesn't contain image mentioned in html page.


ThePavolC:
I was just going through "Forking project guide"

PR #6298, opened 2015-06-23
Simple changes

ecornel:
I've modified the HTML and CSS files a little.


ecornel:
It's my first deal with Git. Thanks.

PR #6407, opened 2015-07-01
Added HelloWorld Project

ShuanWang:
Added a hello world console application. Also provided a user instructions documentation for the hello world application.


ShuanWang:
i think this may be good.

PR #6420, opened 2015-07-03
add 'wangjunting was here ever'

wangjunting:
only add 'wangjunting was here ever' this sentence


gberkwitt:
Learning GitHub. Thanks to all.

PR #6522, opened 2015-07-12
test for commit

Frederico-Takayama:
just a pull request test


Frederico-Takayama:
add a image with forkit.gif name because index.html was with broken image

Frederico-Takayama:
the imported image is not a gitHub image, so i expected a denied pull request

PR #6693, opened 2015-07-29
Add image not found text for forkit.gif

dtaillard:
I set the alt attribute for the forkit.gif image in index.html to show image not found text.


dtaillard:
I created this pull request purely for practice.

PR #6743, opened 2015-08-03
Ravibranch

ravisinghai:
Added some simple changes.

### DIFF
--- a/test1.txt
+++ b/test1.txt
@@ -1,4 +1,8 @@
 This is the first version of the first file.
 
+I have added this new "code" based on the comments / feedback received for the pull request.
+Line 2 of the "code".
+
+
 Let's see how it works in Git.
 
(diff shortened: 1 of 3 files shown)


test1.txt
```diff
@@ -0,0 +1,4 @@
+This is the first version of the first file.
+
+Let's see how it works in Git.
+
```
ravisinghai, line 8: Add some more meaningful messages instead of basic stuff.

testdir/test2.txt
```diff
@@ -0,0 +1,5 @@
+This is the first version of second file in a directory.
+
```
ravisinghai, line 2: Provide more details on what you intend to do.

test3.txt
```diff
@@ -0,0 +1,5 @@
+
+Another new file.
+
+This will contain some useful code.
+
```
ravisinghai, line 5: Where is the useful code?

PR #7014, opened 2015-08-25
Update styles.css

maggiesu:
hi


Jinoooo:
This is not a valid change.  It will break the code and not a good practice.

PR #7127, opened 2015-09-03
I made some changes here

aedoler:
Yes


misterdev:
Now it works really good!

PR #7244, opened 2015-09-12
test.txt has been appeared for test purposes

micalan:
This file is created just for testing purposes


micalan:
The second pull request

PR #7493, opened 2015-10-04
Fix 'forkit.gif' source

khreenberg:
This pull request is just practice. It changes the source of the image in `index.html` to an absolute path found using Google Image Search.


khreenberg:
Hello, World?

PR #8374, opened 2015-12-22
Smile

jbgledy:
Added Smile face


jbgledy:
I added a smile face.  Hope you like it.

PR #8470, opened 2016-01-02
Modify_css_width

NaWu-Dev:
Change the css width setting. Just for try. Thanks.


NaWu-Dev:
First try based on the tutorial. Is that possible to try this?

PR #8647, opened 2016-01-15
Make a simple change.

steerlink:


steerlink:
I want to delete steerlink:simple_text_edit

PR #8657, opened 2016-01-16
Changed 50px to 60px

BeeYean-PCC:


BeeYean-PCC:
changing pix from 50 to 60.

PR #8690, opened 2016-01-19
Made a simple change

patrickfust:
Made a very simple change, so I can try to create a pull-request


patrickfust:
Perhaps it should not be like that

PR #8742, opened 2016-01-22
#2158

usernamealreadyis:
verify #2158


usernamealreadyis:
close: #2158 
open: #8742

PR #8791, opened 2016-01-26
Update style.css

ZachCreates:
Added font color "blue" to p.


DrJid:
Can you make it royal blue instead? I heard that color makes you feel like a prince

PR #9445, opened 2016-03-03
Update README.md

cx1991427:


tdixon777:
I truly think  this update to the readme would be a significant contribution to the overall value of the project.

PR #9527, opened 2016-03-13
Modification to text

kfg966:
Only modified the text description to remove offense.


kfg966:
only slight textual modification removing the offense

PR #9566, opened 2016-03-17
Update README.md

keithworrell:
Rather than saying "Hello There", it would be more appropriate to say "Hello Dude".


johanvanderkuijl:
I think that @keithworrell has an excellent point here.

PR #9633, opened 2016-03-23
made changes that will shock the world

nervren:
Instructions:
1. see changes
2. live happily ever after


nervren:
Comment of life and death.

EDIT of resurrection.

PR #9837, opened 2016-04-12
Changed the background from default to gray

ElvinGarcia:


ElvinGarcia:
Changed the background from default to gray

PR #9952, opened 2016-04-24
HTML EDITING

subakun:
Added two paragraphs. 
one with user name
and one with a table


subakun:
Added two paragraphs to make it look better ;-)

PR #10167, opened 2016-05-16
Brighten things up a bit

ehdevries:
:octocat: Here you go!


ehdevries:
Ahhhh okay. never mind.

PR #10275, opened 2016-05-27
some paragraph

michal-kk:
oh whatever


index.html
```diff
@@ -10,8 +10,9 @@
 <body>
 
 <img src="forkit.gif" id="octocat" alt="" />
-
-<!-- Feel free to change this text here -->
+<p>
+  Giant robots
```
michal-kk: Should be something more

PR #10308, opened 2016-05-31
Changed text in <p> tag

shonore:


shonore:
Changed the <p> tag in the text

shonore:
changed the text

PR #10458, opened 2016-06-14
Update README.md

guanlong1989:
It's just text!


guanlong1989:
It's my 1st time test. thanks

PR #10733, opened 2016-07-16
change "hello" to "hi"

kasaar2:


kasaar2:
Seems like a valuable change.

PR #11055, opened 2016-08-24
added sean

puzzledsean:
added sean.txt for tutorial


puzzledsean:
Ready for review for class

PR #11286, opened 2016-09-19
Update styles.css

LokuDassanayake:
added height


LokuDassanayake:
testing pull function

LokuDassanayake:
teethe

PR #11441, opened 2016-10-07
Trying to make this work.

MartinCernohorsky:
It kind a suck that I can't access files right from GitHub destkop app.

Hmm it allows me to add reactions on my own commetns. Like marking myself with psycho sign.


MartinCernohorsky:
So I can also leave a comment afterwards. Cool.

PR #11575, opened 2016-10-24
add the octocat image

leiguoyang:
@octocat  I find that there's a `img` elemtent in the index.html file and the src attribute is set to forkit.gif, and the gif file is missing in this repository.

So I decide to add the octocat image to this repository. As a result, the webpage looks like this one [demo](http://leiguoyang.github.io/Spoon-Knife).  Does it look nice?


leiguoyang:
Awesome! Approved to merge. 😄

PR #11696, opened 2016-11-02
Changed  `color` for `p` rule.

FilipeBeck:


styles.css
```diff
@@ -14,4 +14,5 @@ p {
   width: 400px;
   margin: 50px auto;
   font: 30px Monaco,"Courier New","DejaVu Sans Mono","Bitstream Vera Sans Mono",monospace;
```
FilipeBeck, line 16: Mude para isso:
```suggestion
  margin: 50px auto;
  font: 40px Monaco,"Courier New","Bitstream Vera Sans Mono",monospace;
```

PR #11716, opened 2016-11-04
ADD A PROVERB

MansaReddy:


index.html
```diff
@@ -11,8 +11,8 @@
 
 <img src="forkit.gif" id="octocat" alt="" />
 
-<!-- Feel free to change this text here -->
 <p>
+  When the going gets tough, the tough get going."
```
leiguoyang: @MansaReddy, I would like to make a tiny modification here, get => gets. Is that right? :-)

PR #11746, opened 2016-11-10
Update index.html

ihashmatullah:
I just changed the comment section only for practice purpose :)


ihashmatullah:
github is kinda fun haha

PR #11835, opened 2016-11-19
fix a typo in README.md

leiguoyang:
I would like to fix a typo from `"Forking Projects""` to `"Forking Projects"`. This is the only one I can find. :smile:


leiguoyang (commented):
thanks for your feedback 👍

PR #11969, opened 2016-12-10
added a smiley face

yurilev:
@octocat, I think it looks better that way :smiley:

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
-  Fork me? Fork you, @octocat!
+  Fork me? Fork you, @octocat! :)
```
leiguoyang, line 16: Awesome :thumbsup:

PR #12178, opened 2017-01-13
Added h1

m-deluca:
addded h1 element inside of body tag


m-deluca:
Added h2 as well inside of the body tag.

PR #12868, opened 2017-04-04
Change index file and added img files

ajm62:
The image file was nowhere to be seen. I added one and made changes to the index files and the style sheet.


ajm62:
Made changes

PR #12938, opened 2017-04-16
Added the absolute image path

millefalcon:


millefalcon:
Well, the image is not viewable, but we know it's an image :}

PR #13255, opened 2017-05-26
Update index

mmosco04:
The image was broken, that is why a new one was added.
Also the text was updated.


mmosco04:
Hello, test

PR #13307, opened 2017-06-02
Consolas

petermenegay:
Consolas is a better font.


petermenegay:
Consolas is a better font. Easier to see.

PR #13453, opened 2017-06-24
Update to index.html

ZZR3556:
Added  message.


ZZR3556:
A friendly message in the index page would be nice.

PR #13487, opened 2017-06-28
Added Rows and Navigation Bars

arismedia:
Added rows and navigation bar


arismedia:
I have tried understanding pull requests. let me know if i did well!

PR #13650, opened 2017-07-24
"xx" is magical sometimes :D

fuhao9202:


fuhao9202:
get ready for next test

PR #13908, opened 2017-08-23
Added fishes and geese

### LINKED ISSUE #13909: Test issue which should be referenced in a pull request
Let's see if the automatic linking works!

rensbaardman:
and merged branch 'change-the-title'

test: this should resolve #13909

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-### Well hello there!
+### The Spoon-Knife Repository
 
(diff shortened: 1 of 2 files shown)


rensbaardman:
this should fix #13909 

edit: see, this doesn't work! The 'magic words' have to be in the original commit description or title; comments don't work that way.

PR #14118, opened 2017-09-21
Removed the word Well from the first phrase

ivokoga:
Just removed the word Well from the first phrase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-### Well hello there!
+### Hello there!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-### Well hello there!
+### Hello there!
```
ivokoga, line 1: That´s okay in my opinion....

ivokoga: cool!

PR #14303, opened 2017-10-17
added test file

keipertk:
test


test.txt
```diff
@@ -0,0 +1 @@
+test
```
keipertk, line 1: This should be changed to test2

keipertk, line 1: this sould be changed to test2

PR #14592, opened 2017-11-21
Fixed up  index.html and styles.css files

lemnok:
Modified index.html and styles.css files


lemnok:
I really think this change is necessary.

PR #14725, opened 2017-12-10
added h2 to index page.

Sayu2003:
Added additional information about octocat.


Sayu2003:
Added h2 as well inside of the body tag.

PR #14728, opened 2017-12-11
Added h1,h2 to index page

nittekume:
added additional information about octocat


nittekume:
Added h1 and h2 inside the body tag

PR #14931, opened 2018-01-09
Just add Hello

jmonclard:
Just add Hello to the tutorial page.
I definitively try to understand how GitHub works. Not really Crystal clear to me, so I try to practice a bit...
My change is really minimal, and you can throw it away !

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,6 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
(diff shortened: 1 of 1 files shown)


jmonclard:
Well perhaps there is really no need to keep this Pull request open !

jmonclard:
So I close it right away.

PR #15365, opened 2018-03-08
Changed text

zarembsky:
Just changed the text


zarembsky:
I like this text more.

PR #15371, opened 2018-03-09
Update styles.css

wmoukhaiber:


wmoukhaiber:
changed width from 400 to 500

PR #15590, opened 2018-04-09
patch 1 - test

andrejkurusiov:
### minor changes to readme
some line of text was added as a test.


andrejkurusiov:
`testing` is done, close #15590 .

PR #16089, opened 2018-06-30
Update to hello world java

anirudh-mathur05:
public class HelloWorldExample
 
 {
 
  public static void main(String args[])
 
  {
 
    System.out.println("Hello World !");

    /* first comment*/

    System.out.println(" second String");
  
  }
 
}


rockstar-rm:
NA

PR #16161, opened 2018-07-15
Added alt text to image and a header.

phajjar:
Trying to commit stuff without fubaring it.


phajjar:
Obviously this isn't going anywhere, but thanks for the nice tutorial.

PR #16187, opened 2018-07-19
Edited the text of index.html for better usability

m-laengsfeld:
What do you think about it? Makes sense, right?


m-laengsfeld:
This is pretty **smart** ...

PR #16255, opened 2018-07-29
Update index.html

Sinoptik125:
I add some thext to the index file


Sinoptik125:
It is great. i am trying to do

PR #16619, opened 2018-09-25
modified index and style

L00149524:
title added

style 550px


L00149524:
trying fork for time,

PR #16788, opened 2018-10-21
update index.html

SirMo7md:
added my GitHub username instead of the octet user name


SirMo7md:
just  a comment here to commit the change

PR #16927, opened 2018-11-04
added h2 to index

sabrina53:
Added additonal information about Octocat


sabrina53:
added h2  inside body tag

PR #17103, opened 2018-12-06
Let me guide you!

hinorashi:
Why join the navy if you can be a pirate!

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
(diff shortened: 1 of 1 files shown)


hinorashi:
Let's get home!

index.html
```diff
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
-  Fork me? Fork you, @octocat!
+  Fork me? Fork you, @hinorashi!
```
hinorashi, line 16: Should it be?

PR #17223, opened 2019-01-02
Modified index.html

dfcbt:
I have changed index.html for clarification.


dfcbt:
That is not a good idea.

PR #17255, opened 2019-01-09
Cool changes

ghost:
Self-describing


me-honmono:
not so cool after all ;-)

PR #17288, opened 2019-01-13
Training branch

RodPell:
First pull from changes made in VSCode


RodPell:
First pull from changes made in VSCode

PR #17785, opened 2019-03-29
I have written a very nice suggestion (testing github) ....

pathros:
I am following this tutorial (https://guides.github.com/activities/forking/). So let's see if people like my great and awesome suggestion.

Hoping that my copy will not exist in infamy on the Internet...

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,5 +16,7 @@
   Fork me? Fork you, @octocat!
 </p>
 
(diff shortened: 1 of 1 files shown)


lratz (changes requested):
make another nice suggestion

lratz:
thanks

PR #17796, opened 2019-03-31
feat: spiral-in message

daisuke-f:
We need this cool animation!


daisuke-f:
> It's not spiral. Just rotating.

Anyway, I believe it's coll.

PR #18058, opened 2019-05-22
Update styles.css

rithuik:
Just checking

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -10,6 +10,7 @@
 }
 
 p {
+  
   display: block;
   width: 400px;
   margin: 50px auto;
--- a/styles.css
+++ b/styles.css
@@ -1,6 +1,7 @@
 * {
   margin:0px;
   padding:0px;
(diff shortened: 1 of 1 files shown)


styles.css
```diff
@@ -10,6 +10,7 @@
 }
 
 p {
+
```
lissy93, line 15: Blank lines within a CSS class group is not convention.

lissy93 (changes requested):
Remove blank line on `styles.css:13`

styles.css
```diff
@@ -1,9 +1,11 @@
 * {
   margin:0px;
   padding:0px;
+  
 }
 
 #octocat {
+  dfdfg
```
lissy93, line 8: That is not even valid CSS. You will break Octocat's website 😨

lissy93 (changes requested):
Your changes are invalid

PR #18378, opened 2019-07-29
Added some chnage in the .css file to check the pull request.

dpjoshi:
My first pull request. Please review the changes and include if okay.

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -11,7 +11,7 @@
 
 p {
   display: block;
-  width: 400px;
+  width: 450px; /*Earlier it was 400px*/
   margin: 50px auto;
   font: 30px Monaco,"Courier New","DejaVu Sans Mono","Bitstream Vera Sans Mono",monospace;
 }


dpjoshi (commented):
Okay !This line was changed.

dpjoshi (commented):
This line is not changed.

styles.css
```diff
@@ -11,7 +11,7 @@
 
 p {
   display: block;
-  width: 400px;
+  width: 450px; /*Earlier it was 400px*/
```
dpjoshi, line 14: This is good enough!

dpjoshi: Okay..thanks for reviewing.

styles.css
```diff
@@ -11,7 +11,7 @@
...
   margin: 50px auto;
```
dpjoshi, line 15: Looks like you have not changed this line.

dpjoshi: Yes. not changed.

dpjoshi (commented):
thanks for the review.

PR #18682, opened 2019-09-21
changed the src of the image to a server location

Prottoy2938:
Changed the image to an src link so everyone can see it


Prottoy2938:
This is good

Prottoy2938:
this is good

PR #19281, opened 2019-12-17
へ

morigorikun:
# what
へ
# why
test

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 ### Well hello there!
-### he
+### he he
 
 This repository is meant to provide an example for *forking* a repository on GitHub.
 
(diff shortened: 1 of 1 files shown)


kingclowndean:
https://github.com/octocat/Spoon-Knife/pull/19281.diff

kingclowndean:
an·ec·do·tal /ˌanəkˈdōdl/ adjective  1. (of an account) not necessarily true or reliable, because based on personal accounts rather than facts or research: "while there was much anecdotal evidence there was little hard fact" synonyms informal, unreliable, based on hearsay, unscientific antonyms experimental, scientific

PR #19436, opened 2020-01-19
ADDED COLOR IN THE BACKGROUND

Narrai:
Just added one change in the background


Narrai:
added purple background

Narrai:
adding purple background

PR #19486, opened 2020-01-30
Changing the background and add some underline.

PiotrNap:


sajjadmajid77:
[Variables]
MyVar1=This is a string!
MyVar2=So is this!
Red=255
Green=150
Blue=0
Alpha=90
FB_IMG_1486094653636 (1)
seras_victoria__christmas__by_themaestronoob_ddmyodw

PR #20128, opened 2020-04-25
Attempt 1.0

ghost:
Clicking on Compare and Pull Request sends you to a discussion page, where you can enter a title and optional description. It’s important to provide as much useful information and a rationale for why you’re making this Pull Request in the first place. The project owner needs to be able to determine whether your change is as useful to everyone as you think it is.

When you’re ready typing out your heartfelt argument, click on Send pull request. You’re done!

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,6 +4,7 @@
 <head>
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
   <title>Maluba</title>
+  <p>FOZZANAPOLI<p/>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
-  Fork me? Fork you, @octocat!
+  Fork your dog
```
jhosman, line 17: Test comment created with REST API

PR #20146, opened 2020-04-27
New color

aska0407:
New text with a changed color.


macmaster (approved):
Thanks for the changes.
They look really good.

PR #20301, opened 2020-05-16
just added image and background-color property

singhAmandeep007:
hopefully my pull request gets accepted haahahha..


UbuntuLover1337:
hahaha

looks great!!!

PR #20322, opened 2020-05-19
Index Page Changed

Web5R:
Dear,
thanks for the tutorial.

You might or might not consider my humble contribution..

;-)
Bart

### DIFF
--- a/styles02.css
+++ b/styles02.css
@@ -1,0 +1,17 @@
+* {
+  margin:0px;
+  padding:0px;
+}
+
+#octocat {
+  display: block;
+  width:384px;
+  margin: 50px auto;
+}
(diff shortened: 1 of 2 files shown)


styles02.css
```diff
@@ -0,0 +1,17 @@
...
+  display: block;
+  width:384px;
+  margin: 50px auto;
+}
+
+p {
+  display: block;
+  width: 400px;
+  margin: 50px auto;
+  font: 30px Monaco,"Courier New","DejaVu Sans Mono","Bitstream Vera Sans Mono",monospace;
```
mrkrabs654, line 16: I think comic sans would look better here.

PR #20332, opened 2020-05-20
lolololololol

UbuntuLover1337:
ahahahahah

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -13,7 +13,7 @@
 
 <!-- Feel free to change this text here -->
 <p>
-  Fork me? Fork you, @octocat!
+  Fork me? fuck you, @octocat!
```
macmaster, line 16: would like to know what @octocat  thinks of this change.

PR #20348, opened 2020-05-22
colour change

burnish-studio:
colour changed to aquamarine


styles.css
```diff
@@ -1,6 +1,7 @@
 * {
   margin:0px;
   padding:0px;
+  background-color: aquamarine;
```
chrisjbarnatt, line 4: I think a nice yellow colour would be better here.

PR #20354, opened 2020-05-23
Create simple test.cpp

chrisjbarnatt:
Just some c++ code i was working on over the covid-19 pandemic whilst i was at home being bored. Feedback is welcome and very much appreciated. It was an experiment in using code which was, at the time, not fully tested by researchers. Im kinda ashamed of it but was lucky to have access to free dev lights while i was doing this. thanks!

All I can say is WOW. Would have rated 5 stars had I been able to post before, but.. What a time warp. Well, with hundreds of orders there was no way I could release the units in just a day, so here we are with 3 weeks to wait until we get in. This project is definitely not for everyone. Its not deep. Not math heavy or anything.

### DIFF
--- a/test.cpp
+++ b/test.cpp
@@ -1,0 +1,8 @@
+[[gnu::noinline]] void cool() {
+    cool();;;;;;}
+
+auto 
+ main(   void ) -> decltype(auto) {
+ 
+       cool(    );
+                                                                    }


chrisjbarnatt:
Any reviews/approvals are appreciated.

PR #20611, opened 2020-06-26
docs: Replace @octocat with @karnaker

karnaker:
Testing this test


karnaker:
@octocat Testing comment

PR #20728, opened 2020-07-08
Made a better readme

Lvrks:
As you can see, I made a better readme. Merge if you also think it's better.


BETTERREADME.md
```diff
@@ -0,0 +1 @@
+I THINK YOUR README COULD BE BETTER
```
Lvrks, line 1: Looks good!

PR #21131, opened 2020-08-23
Added a sample text file.

shafeeshafee:
A handy sample file, just in case you need to write something down when working 😄


prkagrawal (commented):
It is empty.

PR #21760, opened 2020-10-24
Update index.html

andre-wells:
Added super-necessary change :D


andre-wells:
I think I've got this. Closing!

PR #22333, opened 2021-01-22
Tutorial edit.

moezilla402:
Just sending a request in the process of going through a tutorial.  Thx!


moezilla402:
Changed text on line 16 as a part of an online tutorial.

PR #22404, opened 2021-02-02
Just added shake.speare.

bashelled:
I just added Shakespeare quotes. That's all.


bashelled:
Well this is old

bashelled:
nah jk

PR #22566, opened 2021-02-20
Removed user name from index.html message

JackNWhite:
In the spirit of general misanthropy, it's better not to insult octocat in particular, and instead to insult everyone.


aleee1333 (commented):
Ok

PR #22615, opened 2021-02-26
Rename index.html to ximena.html

ximenahan:
remane index.html to ximena.html. Just the practice following the instruction


yra40:
Privet

PR #22911, opened 2021-04-04
Style edits

WombleWoo7547:
# Style edits

* [x] Separate ``` font ``` style to be in ``` font-size ``` and ``` font-family ```
* [x] Change ``` font-family ``` to ``` sans-serif ``` rather than ``` Monaco, "Courier New", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", monospace ```

### DIFF
--- a/styles.css
+++ b/styles.css
@@ -13,5 +13,6 @@ p {
   display: block;
   width: 400px;
   margin: 50px auto;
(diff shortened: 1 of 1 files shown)


WombleWoo7547:
Something very simple
```

I like code

```

PR #23023, opened 2021-04-20
Shortening text

FabsOliveira:


UNIVAC-Coder (commented):
I (sort of) like your suggested changes !

PR #23322, opened 2021-05-26
make everything purple

shahzadqadir:
background colour is changed to purple.


shahzadqadir:
changed background colour to purple.

PR #23563, opened 2021-07-06
Change css

mmatsumoto1026:


mmatsumoto1026:
OK. Text's color in p tag has changed.

PR #23596, opened 2021-07-11
Update index.html

PanCzaro:


PanCzaro:
I was following instruction

PR #23642, opened 2021-07-22
Update content in the file

ivansing:
Changed title and add some description to the index file.


ivansing:
This was a helful update for the index page.

PR #23934, opened 2021-09-16
Updated html and css

JPMurara:
1. html - changed the paragraph
2. css - added image


JPMurara:
Thanks for creating a test repo! Really helps to practice

PR #24032, opened 2021-10-08
Inital creation of Additional Instructions.txt

ToastNGhost:


ToastNGhost:
Why do users need additional instruction though? I think the ones on the guides page are enough.

PR #24266, opened 2021-11-15
Update index.html with rejected reference

memcpy-rand-rand-rand:
Single line added to existing paragraph.


KaliberXout:
https://github.com/octocat/Spoon-Knife/pull/24266#issue-1054037384

PR #24596, opened 2022-01-15
added a alt text in image in index.html

SpectralGT:
I added a alt text in img in index.html
It will be useful if the image don't load , then the alt text will be shown.

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,7 +9,7 @@
 
 <body>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -9,7 +9,7 @@
 
 <body>
 
-<img src="forkit.gif" id="octocat" alt="" />
```
maulibrhm1, line 12: kesalahan

PR #24688, opened 2022-01-30
changed width of octocat from 384 px to 385 px

sushskvnitn:
The width of the octacat has gone from 384 px to 385 px.


gibbiemonster (approved):
Seems fine.

PR #24762, opened 2022-02-05
Update styles.css

kev-web:
changed the body color to pink.


kev-web:
Again replace the background-color from pink to cyan

kev-web:
another comment here...form pull request of mine.

PR #26453, opened 2022-03-05
So I had this crazy idea..

Atinon:
Why not just make it all purple?!
also shoutout to https://www.youtube.com/c/ColtSteeleCode for the GitHub guide!


Atinon:
Purple.

PR #27406, opened 2022-07-20
changed a bit of text in index

unfirthman:
deleted something about "fork you, octocat!" and replaced with new text


cruZitoc03:
> deleted something about "fork you, octocat!" and replaced with new text

PR #28584, opened 2022-11-23
World cup part 2

gambigivesyouwiings:
Japan beat Germany? Oh my, Group E may well turn out to be the group of death


66667888sssttt:
> 

💶✅

PR #29900, opened 2023-04-27
Better late than never...

chrissuozzo:
🎉


README.md
```diff
@@ -6,4 +6,4 @@ Creating a *fork* is producing a personal copy of someone else's project. Forks
 
 After forking this repository, you can make some changes to the project, and submit [a Pull Request](https://github.com/octocat/Spoon-Knife/pulls) as practice.
```
sboddu2, line 8: # Don't put loop here

PR #30461, opened 2023-07-17
Update index.html

theswiftvlad:
Test comment)

### DIFF
--- a/index.html
+++ b/index.html
@@ -12,6 +12,7 @@
 <img src="forkit.gif" id="octocat" alt="" />
 
 <!-- Feel free to change this text here -->
+<p>Hello Octocat!<\p>
 <p>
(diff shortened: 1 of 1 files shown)


stysan:
### why this won't get merged
this repo is already perfect for what it's purpose is and adding a random line of code without a purpose is _kinda weird_
### for next time you want to merge something
please think if something you want to merge is actually needed in the project - because this won't get merged because you added a line of code for no reason! and you can argue a whole lot about whether you need it or not, but you don't actually need your `<p>Hello Octocat!</p>` thing.
### here's why
1. it doesn't have a purpose
2. it is not ideal line of hello world type code, because @octocat is already used
3. it doesn't fit in the style of this code  
  
**thank you for your time!**

PR #30545, opened 2023-08-01
Add a hello image to README.md and index.html

wtrape:
I also changed styles.css to make the html look well.


Toohott23 (changes requested):
Delete

PR #30662, opened 2023-08-15
On branch issue1

segnbi:
modified:   index.html
	  -> added title "h1" and changed "img" id to segnbi.


segnbi:
I found out the html file didn't have a title

PR #31713, opened 2024-01-12
Additions to index.html + style.css

ebutbetter:
Added a table with bread things to index.html, styled it in style.css


rockundeads (changes requested):
Difference memory request Info customer

PR #32366, opened 2024-02-22
Update index.html

HarshithVarmaG:
to include my  GitHub username "HarshithVarmaG" in HTML file using the dynamically insert username using JavaScript approach


HarshithVarmaG:
To include my GitHub username in HTML file using the dynamically insert username using JavaScript approach.

PR #33487, opened 2024-07-25
viewport meta added

vitaly-belov:
The browser's [viewport](https://developer.mozilla.org/en-US/docs/Glossary/Viewport) is the area of the window in which web content can be seen. This is often not the same size as the rendered page, in which case the browser provides scrollbars for the user to scroll around and access all the content.


vitaly-belov:
it was educational

PR #33525, opened 2024-07-31

VadymKr:
Addressing #1


VadymKr:

PR #33830, opened 2024-09-12
Add personal message to index and style files

kestrel-123:
As the idea of this lesson file is to simply add some personal alterations, I added a div with an id of
"kestrel-special" and wrote a meassage/signature
using a nested paragraph element inside the div.
I then restyled my kestrel-special div using the style sheet, including a hover effect just for fun.

### DIFF
--- a/index.html
+++ b/index.html
@@ -16,5 +16,9 @@
   Fork me? Fork you, @octocat!
 </p>
 
+<div id="kestrel-special">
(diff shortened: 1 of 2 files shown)


kestrel-123:
it's <em>nice</em> to be able to <b>test</b> <mark>these things</mark> out :sparkles:

PR #33960, opened 2024-09-27
test with empty file

josefuchs:


Pricer187 (commented):
How do you feel about this is it good or what I'm asking you for your honest opinions I don't want to make a bad choice anymore

PR #34581, opened 2024-11-30

JT69C10:
Addressing #1


JT69C10:
Not sure if this is going to work or not

PR #35246, opened 2025-02-04
Made changes

Sathvik0806:


Sathvik0806:
i changed some part of code of spoon_knife

PR #35270, opened 2025-02-05
1st Update index.html

T-Karthik-Reddy:
Changed the text in paragraph element and now pulling from BombSpoon branch


T-Karthik-Reddy:
done

PR #35573, opened 2025-02-19

nmilety:
Addressing #1


nmilety:
Just following Microsoft Learning Path instructions

PR #35955, opened 2025-03-27

amoremax092:
Addressing #1


Amjadkkh:
Recovery and update with my finance app

PR #37116, opened 2025-07-07

ensotech:
Addressing #1

### DIFF
--- a/index.html
+++ b/index.html
@@ -9,12 +9,13 @@
 
 <body>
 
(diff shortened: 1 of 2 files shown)


ensotech:
.

ccharmed17 (commented):
I don't think I'm collaborating with the right people and honestly I don't get it I'm just now starting IT work in my college classes. I didn't pick this group of people and I don't want to be cloned so can you fix everything that is wrong on here

PR #37498, opened 2025-08-12
ajout app.html

JulienNeville:
add a html file


ssmsmsmsmm:
FIRST LOCATION - DAY

Some scene description, using the Action element. ["INT. FIRST LOCATION - DAY", above, is a Scene Heading element.]

The FIRST CHARACTER strolls in.

FIRST CHARACTER

Even though I'm alone, I can talk to myself with a Dialogue element.

A SECOND CHARACTER comes in, joining the First Character.

SECOND CHARACTER

(whispering)

And if you need to be a little more specific about how you're speaking, that's what the Parenthetical element is for.

CUT TO:

EXT. SECOND LOCATION - NIGHT

The First Character is alone again here. [It's not normally necessary to indicate transitions like "CUT TO:", above, but if you really need to, that's what the Transition element is for.]

FIRST CHARACTER

Wait, how did I get here now?

ANGLE ON

a screenplay in the First Character's hands

PR #37775, opened 2025-09-12
taco commit to html

premgallant:
add ignore and modify html file


premgallant:
looks good to me. merging to main

PR #37813, opened 2025-09-18
Updated the html and css

MaryO-droid:


MaryO-droid:
This pull request includes changes to the html and css

PR #37965, opened 2025-10-02

ghost:


furkan19899-hash:
httpAsking for help, clarification, or responding to other answers. Making
statements based on opinion; back them up with references or personal
experience. To learn more, see our tips on writingstrong text scs scs\n"

is normal text. 2. So is this, but now follows a code block:

anowardear062-svg (approved):

PR #37977, opened 2025-10-04
add a line to the file README

denisok37:
ONEPIECE THIS IS REAL


pNm193 (changes requested):
Needs Rules for vulnerability and security to be set. Ruleset for no unverified. Ruleset for no illegal content. Ruleset for no dependabots or any bots.

PR #38106, opened 2025-10-15
Uptdate index.html and styles.css

bypbyp:
Addressing #1


bypbyp (commented):
updated index.html and styles.css

PR #39699, opened 2026-04-06

DAMoorhouse:
Addressing #1

Update collaboration message and styles in index.html


DAMoorhouse:

PR #39946, opened 2026-05-06
changed text colour to red

diamond-k:
Changed text colour to red


Octocatking7094:
Looks good! The text color change to red has been implemented correctly. Approved ✅Thanks for the contribution 👍 Just to confirm, was the text color change applied to all text or only specific parts? Could you clarify?

diamond-k:
The change applies to `<p>` text only.

PR #39968, opened 2026-05-10
added h1 to the html file

nalsvanessa:


Octocatking7094:
Nice and simple! ✅  
Good job adding the h1 to the HTML file. Solid practice PR!

PR #40148, opened 2026-06-06
ajout fichier

Akashi-225:
ce fichier a pour de tester la fonction fork

### DIFF
--- a/test
+++ b/test
@@ -1,0 +1,1 @@
+juste un test


christian03kalu-art (commented):
Text review @tygabillionz

christian03kalu-art (approved):

christian03kalu-art:
#include <stdio.h>
#include <unistd.h>
#include <string.h>
#include <sys/types.h>
#include <sys/wait.h>

int main() {
    // Array to hold pipe file descriptors:
    // pipefd[0] is the READ end
    // pipefd[1] is the WRITE end
    int pipefd[2]; 
    pid_t pid;
    char buffer[100];

    // 1. Create the pipe BEFORE the fork
    if (pipe(pipefd) == -1) {
        perror("Pipe creation failed");
        return 1;
    }

    pid = fork();

    if (pid < 0) {
        perror("Fork failed");
        return 1;
    } 
    else if (pid == 0) {
        // --- CHILD PROCESS (Sender) ---
        // The child only WRITES, so close the unused READ end
        close(pipefd[0]); 

        char message[] = "Secret message from the Child process!";
        printf("[Child] Sending data through the pipe...\n");

        // Write the message string (including the null terminator) to the pipe
        write(pipefd[1], message, strlen(message) + 1);

        // Close the write end when finished
        close(pipefd[1]); 
        return 0;
    } 
    else {
        // --- PARENT PROCESS (Receiver) ---
        // The parent only READS, so close the unused WRITE end
        close(pipefd[1]); 

        printf("[Parent] Waiting for data from child...\n");

        // Read data from the pipe into the local buffer
        // This blocks execution until the child writes data
        read(pipefd[0], buffer, sizeof(buffer));

        printf("[Parent] Received message: \"%s\"\n", buffer);

        // Close the read end when finished
        close(pipefd[0]); 

        // Clean up the child process
        wait(NULL); 
    } Transaction Receipt-260605010100102583981468.pdf



    return 0;
}

PR #40755, opened 2026-08-04

Vaakii22:


liu6381810 (commented):
Quick pass on the copy refresh — one inline note.

Issue #1, opened 2011-02-25
Needs moar pun


octocat:
Seriously. Not punny enough.

kevasdf:
This isn't done, so don't stick a fork in it yet.

ericchen0121:
Waiting for those puns. Fork 'em up.

backwoodsman:
```
for($k = 4;$k < $tuning; $k++){
    $pitch = $fork;
```

mayaralinekell:
oi

chrisyco:
There will _never_ be enough pun!

carbonatedcaffeine:
PUN... heh eheh

Nyameliaaaa:
Boy Looks Like Some Good Forking Issue

SwiftyProgrammer690:
ummm

silverkorn:
Wow!

kingclowndean:
Wow we

EmmmaTech:
This issue needs more puns

SwiftyProgrammer690:
ya need more punnn

Amywalkerpowell:
Never enough punn

goilat:
I have not hacked anybody....Hope you either....please J, focus on ur
business.

CybCom:
open it! 😄

Rockyspade:
> Seriously. Not punny enough.

> needs more moar pun #34233 #32693 [[]](https://github.com/octocat/Spoon-Knife/issues/1#issue-628378)(url)

SwiftyProgrammer690:
why is this still alive?

alexanderregnier:
> Esperando esos juegos de palabras. ¡Que les vaya bien!

SwiftyProgrammer690:
Why is this still alive

Debelopercgx:
> So can this code me 8nto rec I eving a d9nation from one bank account to my savings as well. Or just unimag8ned ideals

trangiabach:
Great issue!

trangiabach:
Great issue!

AW1534:
i think its quite punny

Clexus:
hello

junaid-shirur:
example_body

junaid-shirur:
example_body

junaid-shirur:
example_body

junaid-shirur:
example_body

junaid-shirur:
example_body

junaid-shirur:
example_body

[...]

PR #41131, opened 2026-08-30
Add practice comment to index.html

devtesu:


devtesu:
This is a practice PR; it references octocat/Spoon-Knife#1 as an example of a cross-reference.

PR #41133, opened 2026-08-31
Add note about octocat's disappearance

krwrks:
Added a note about octocat's whereabouts.


krwrks (commented):
I wish I could approve for you, but alas!

PR #41354, opened 2026-09-15

Krishna-krishnan:
Addressing #1


index.html
```diff
@@ -9,11 +9,12 @@
```
Sasha84-code: https://git.github.io/htmldocs/git-worktree.html

